PR #50, opened 2021-03-05
Fix versioning issues

dionhaefner:
Investigating #49 

Tweaks:
- Move runtime dependencies to `setup.py`. There is no benefit in specifying those twice, and if only `setup.py` is universally supported, so be it.
- Ensure that `setup.py` is aware of `setuptools-scm`.
- Use `setuptools-scm-git-archive` to inject version into GitHub release tarballs.
- Bump minimum setuptools version [as recommended](https://github.com/pypa/setuptools_scm/#pyprojecttoml-usage).

### DIFF
--- a/.git-archival.txt
+++ b/.git-archival.txt
@@ -1,0 +1,1 @@
+ref-names: $Format:%D$
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+.git_archival.txt  export-subst
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,3 @@
-[tool.setuptools_scm]
-write_to = "mpi4jax/_version.py"
-
-
 [tool.coverage.run]
 branch = true
 parallel = true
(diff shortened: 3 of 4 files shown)


PhilipVinc:
I think mpi4py must stay in the toml because it’s a build time dependency, and we need it to run setup.py.

PhilipVinc:
@dionhaefner this should have fixed the problem?

dionhaefner:
I think so, at least the error should be different now. This requires `setuptools-scm` to be present whenever `setup.py` is invoked.

PhilipVinc:
By running
```
➜ pip install -U setuptools setuptools_scm[toml] wheel twine mpi4py cython
Python setup.py sdist
```

And inspecting the resulting source wheel, I’m unsure that it is resolved.
The setup.cfg file contains a setup.cfg with unpopulated [egg_info]tag_build

Il 5 mar 2021, 16:28 +0100, Dion Häfner <notifications@github.com>, ha scritto:

dionhaefner:
That shouldn't matter. The `setuptools-scm` docs say 

> If you see unusual version numbers for packages but python setup.py --version reports the expected version number, ensure [egg_info] is not defined in setup.cfg.

So it sounds like an empty `tag_build` is expected. You can see that the correct version is injected in `PKG-INFO` and `mpi4jax/_version.py`.

I noticed that you can still run `setup.py` install without having `setuptools-scm` installed, though. I'll revisit that to make sure an exception is raised instead.

PR #55, opened 2021-03-09
Implement missing primitives

### LINKED ISSUE #1: MPI primitives
This is a list of everything that's supported by `mpi4py`:

- Abort
- Accept
- Allgather
- Allgatherv
- Allreduce
- Alltoall
- Alltoallv
- Alltoallw
- Barrier
- Bcast
- Bsend
- Bsend_init
- Call_errhandler
- Cart_map
- Clone
- Compare
- Connect
- Create
- Create_cart
- Create_dist_graph
- Create_dist_graph_adjacent
- Create_graph
- Create_group
- Create_intercomm
- Create_keyval
- Delete_attr
- Disconnect
- Dup
- Dup_with_info
- Exscan
- Free
- Free_keyval
- Gather
- Gatherv
- Get_attr
- Get_errhandler
- Get_group
- Get_info
- Get_name
- Get_parent
- Get_rank
- Get_size
- Get_topology
- Graph_map
- Iallgather
- Iallgatherv
- Iallreduce
- Ialltoall
- Ialltoallv
- Ialltoallw
- Ibarrier
- Ibcast
- Ibsend
- Idup
- Iexscan
- Igather
- Igatherv
- Improbe
- Iprobe
- Irecv
- Ireduce
- Ireduce_scatter
- Ireduce_scatter_block
- Irsend
- Is_inter
- Is_intra
- Iscan
- Iscatter
- Iscatterv
- Isend
- Issend
- Join
- Mprobe
- Probe
- Recv
- Recv_init
- Reduce
- Reduce_scatter
- Reduce_scatter_block
- Rsend
- Rsend_init
- Scan
- Scatter
- Scatterv
- Send
- Send_init
- Sendrecv
- Sendrecv_replace
- Set_attr
- Set_errhandler
- Set_info
- Set_name
- Spawn
- Spawn_multiple
- Split
- Split_type
- Ssend
- Ssend_init

(plus lower-case variants to communicate Python objects)

#### Thoughts:

- Non-blocking operations (e.g. `Isend`) are perhaps out of scope for now, since I don't see a way to transparently hand back the request object to Python. I guess you could do it C-style and hand back a raw pointer if you really want to support non-blocking ops.
- Lower-case operations (e.g. `send`) seem cumbersome to implement and not that useful from JITed code.

I would suggest the following minimal list to start out with:

- Abort
- Allgather
- Allreduce
- Alltoall
- Barrier
- Bcast
- Disconnect
- Gather
- Get_rank
- Get_size
- Recv
- Reduce
- Scan
- Scatter
- Send
- Sendrecv
- Spawn
- Split

I could probably contribute some of those if you agree.

dionhaefner:
Fixes #1

To-do:
- [x] MPI wrappers
- [x] CPU implementation
- [x] GPU implementation
- [x] JAX primitives
- [x] Add to docs
- [x] Tests
- [x] Run tests on GPU
- [x] Check docs


dionhaefner:
@PhilipVinc I need to make a decision and could use your opinion.

This regards primitives like `bcast` and `gather` that behave differently depending on whether it is run on the root process or not.

For example, `bcast` output only makes sense in non-root processes that receive the data. `gather` output only makes sense on the root as all others are just sending data over.

Let's use `gather` as an example. I see 4 different ways of handling this:

1. Have `gather` return data and token on root, and only the token on non-root. I.e.,
   
   ```python
   if rank == root:
       data, token = mpi4jax.gather(x, root=root)
   else:
       token = mpi4jax.gather(x, root=root)
   ```

2. Return full data on root process and unmodified input on non-root processed. This means that the return argument has different shapes depending on the process, `(size, *shape)` on root and `shape` on non-root.
3. Return data of same shape on all processes, but data contains uninitialized memory on non-root.
4. Return data of same shape on all processes, but data contains zeros on non-root. (A bit awkward to implement on GPU.)

What do you think? `bcast` is currently using approach 2 which is less awkward there because the shapes are the same.

dionhaefner:
For the record, `mpi4py` returns `None` on non-root.

PhilipVinc:
So.. for `gather` I think option 1 is the most sense full, indeed.

However, for bradcast `bcast` I think it makes sense to output the array in all ranks so that you can write something like

```python
def mpi_split(key, root=0, comm=MPI.COMM_WORLD) -> PRNGKeyType:
    keys = jax.random.split(key, n_nodes)
    keys, _ = mpi4jax.bcast(keys, root=root, comm=comm)
    return keys[comm.Get_rank()]
 ```

which to me seems nicer?

**Comment by dionhaefner** [comment]
Yes, agreed. I think `gather` is really a special case, all the other primitives generate data of the same shape in all ranks. I'll just return the input to be consistent.

**Comment by dionhaefner** [comment]
Sorry, I ripped out the deprecation already because I couldn't be bothered to add them for the new primitives. If you really still want them I can add it back in.

**Comment by PhilipVinc** [comment]
Just leave the ones that are there, there.

There’s no need to deprecate something that was never there, so the new functions don’t need ‘em.

I need the deprecations because we don’t upper bound mpi4jax in netket...
Il 10 mar 2021, 19:54 +0100, Dion Häfner <notifications@github.com>, ha scritto:

**Comment by dionhaefner** [comment]
All of these pass on my test machine:

```bash
$ JAX_PLATFORM_NAME=cpu pytest .
$ JAX_PLATFORM_NAME=cpu mpirun -n 4 pytest .
$ JAX_PLATFORM_NAME=gpu MPI4JAX_USE_CUDA_MPI=0 pytest .
$ JAX_PLATFORM_NAME=gpu MPI4JAX_USE_CUDA_MPI=0 mpirun -n 4 pytest .
$ JAX_PLATFORM_NAME=gpu MPI4JAX_USE_CUDA_MPI=1 pytest .
$ JAX_PLATFORM_NAME=gpu MPI4JAX_USE_CUDA_MPI=1 mpirun -n 4 pytest .
```

The XLA bridge is a mess, especially on GPU, but hopefully we won't have to touch it in the future.

mpi4jax/_src/collective_ops/allgather.py
```diff
@@ -0,0 +1,178 @@
...
+        ]
+    )
+
+    descriptor = build_allgather_descriptor(
+        _send_nitems,
+        _send_dtype_ptr,
+        # we only support matching input and output arrays
+        _send_nitems,
+        _send_dtype_ptr,
+        #
```
PhilipVinc, line 144:
```suggestion
```

mpi4jax/_src/collective_ops/alltoall.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+
+    operands = (
+        _constant_s32_scalar(c, _nitems_per_proc),
+        x,
+        _constant_u64_scalar(c, _dtype_ptr),
+        # we only support matching input and output arrays
+        _constant_s32_scalar(c, _nitems_per_proc),
+        _constant_u64_scalar(c, _dtype_ptr),
+        #
```
PhilipVinc, line 100:
```suggestion
```

PhilipVinc (approved):

mpi4jax/_src/collective_ops/allgather.py
PhilipVinc: are those here for a reason?

dionhaefner: They delimit the comment above :)

PR #69, opened 2021-03-17
Add action to check if GPU extensions can be built

### LINKED ISSUE #44: Check that the GPU version builds (At least)

dionhaefner:
Fixes #44


dionhaefner:
No idea why the OSX tests are failing, maybe a caching issue.

I know you're working on proper GPU CI, but in the meantime this shouldn't hurt.

dionhaefner:
[jaxlib is broken on OSX at the moment](https://github.com/google/jax/issues/6111). So we can just wait this out.

PR #155, opened 2022-07-29
Handle Cray MPICH wrapper compilers

felker:
See #153.


PhilipVinc (approved):
looks good to me. @dionhaefner what do you think?

dionhaefner:
I wonder if at this point it wouldn't be easier to just use the appropriate compiler wrappers as returned by `mpi4py` rather than trying to parse the flags. I can play around with it a bit.

dionhaefner:
Superseded by #157.

PR #189, opened 2023-02-16
Investigate #188

dionhaefner:


dionhaefner:
Merging this now; don't know if it fixed the problem but this shouldn't hurt.

Issue #256, opened 2024-09-13
shallow_water example: TypeError: 'TokenType' object is not iterable


benkirk:
I'm attempting to run  the shallow water benchmark with the following host environment:
```pre
jax                0.4.31
jax-cuda12-pjrt    0.4.31
jax-cuda12-plugin  0.4.31
jaxlib             0.4.31
mpi4jax            0.5.3
mpi4py             4.0.0
```
and receive this error message:
```pre
Traceback (most recent call last):
  File "/glade/derecho/scratch/benkirk/conda-recipes/test_envs/./tests/mpi4jax_shallow_water.py", line 584, in <module>
    sol = solve_shallow_water(t1=10 * DAY_IN_SECONDS, num_multisteps=PLOT_EVERY)
  File "/glade/derecho/scratch/benkirk/conda-recipes/test_envs/./tests/mpi4jax_shallow_water.py", line 428, in solve_shallow_water
    h, u, v = get_initial_conditions()
  File "/glade/derecho/scratch/benkirk/conda-recipes/test_envs/./tests/mpi4jax_shallow_water.py", line 168, in get_initial_conditions
    h0_local, token = enforce_boundaries(h0_local, "h", token)
  File "/glade/derecho/scratch/benkirk/conda-recipes/test_envs/./tests/mpi4jax_shallow_water.py", line 255, in enforce_boundaries
Traceback (most recent call last):
  File "/glade/derecho/scratch/benkirk/conda-recipes/test_envs/./tests/mpi4jax_shallow_water.py", line 584, in <module>
    token = mpi4jax.send(send_arr, dest=send_proc, comm=mpi_comm, token=token)
  File "/glade/derecho/scratch/benkirk/conda-recipes/test_envs/mpi4jax/lib/python3.11/site-packages/mpi4jax/_src/validation.py", line 90, in wrapped
    return function(*args, **kwargs)
  File "/glade/derecho/scratch/benkirk/conda-recipes/test_envs/mpi4jax/lib/python3.11/site-packages/mpi4jax/_src/collective_ops/send.py", line 75, in send
    sol = solve_shallow_water(t1=10 * DAY_IN_SECONDS, num_multisteps=PLOT_EVERY)
  File "/glade/derecho/scratch/benkirk/conda-recipes/test_envs/./tests/mpi4jax_shallow_water.py", line 428, in solve_shallow_water
    h, u, v = get_initial_conditions()
[...]

PR #259, opened 2024-09-17
Fix docs

dionhaefner:
See #256


dionhaefner:
Let's see if that does the trick.

PR #279, opened 2025-03-18
Remove token mode

dionhaefner:


PhilipVinc (approved):
Seems good, thanks!

We should bump the minor version to 0.8 with this to signal the breaking change.
(Also, we could keep a mention that older versions of mpi4jax had tokens but those should not be removed and are no longer returned/accepted, but not sure how important it is).

